PR #62, opened 2018-06-20
APL-75 (Fixed admin password bug)

GAsdeX:


0486:
make password input form as closable modal, no as page part.

PR #75, opened 2018-06-21
Added auto login functionality

0486:


0486:
everything else working file,

PR #77, opened 2018-06-22
Apl 63 to testnet

0486:


0486:
API working, but can not compile tests

PR #153, opened 2018-08-14
Fix decripting of messages

GAsdeX:


0486:
Tested, reviewed, working.

PR #192, opened 2018-09-20
Apl 380: Splash preloader replace...

0486:


src/main/java/com/apollocurrency/aplwallet/apldesktop/DesktopApplication.java
```diff
@@ -226,18 +226,25 @@ public void startAndShow() {
...
+            AnchorPane.setTopAnchor(versionText, 130.0);
+            AnchorPane.setLeftAnchor(versionText, 190.0);
+            AnchorPane.setTopAnchor(statusText, 200.0);
             AnchorPane.setLeftAnchor(statusText, 60.0);
             pane.getChildren().add(statusText);
+            pane.getChildren().add(versionText);
             screenStage.setScene(scene);
-            screenStage.setHeight(250);
-            screenStage.setWidth(400);
+            screenStage.setHeight(217);
```
AndrewBoyarsky: use screenStage.sizeToScene() instead of copying size from css.

src/main/resources/css/splash-screen.css
```diff
@@ -14,6 +14,12 @@
     -fx-fill: white;
     -fx-font-size: 15;
 }
+
+#version-text {
```
AndrewBoyarsky: instead of styles duplication use general class for white text with size 15

PR #202, opened 2018-10-04
APL-388 - 2FA impl(APL-385) + new accounts support + bugfix

AndrewBoyarsky:


Digital21:
The project is a big one and management is taking active action to improve the efficiency of the project.
I hope it will be a great success and to be an excellent potential project in the near future.

PR #232, opened 2018-11-12
APL-465 - Rename attachment, block, currency, transaction variables and db columns

AndrewBoyarsky:
Keep backward compatibility


AndrewBoyarsky:
Warning! Fork will occurred after this code acceptance.

PR #240, opened 2018-11-23
APL-468 Integrate new UI into desktop wallet

leonidstr:
- Add Error Handler

### DIFF
--- a/src/main/java/com/apollocurrency/aplwallet/apl/http/APIErrorHandler.java
+++ b/src/main/java/com/apollocurrency/aplwallet/apl/http/APIErrorHandler.java
@@ -1,0 +1,32 @@
+package com.apollocurrency.aplwallet.apl.http;
+
+import com.apollocurrency.aplwallet.apl.Apl;
+import org.apache.commons.io.IOUtils;
+import org.eclipse.jetty.servlet.ErrorPageErrorHandler;
+
+import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


src/main/java/com/apollocurrency/aplwallet/apl/http/APIErrorHandler.java
```diff
@@ -0,0 +1,32 @@
...
+    public void handle(String target, org.eclipse.jetty.server.Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException {
+        if(response.getStatus() == HttpURLConnection.HTTP_NOT_FOUND){
+            String apiResourceBase = Apl.getStringProperty("apl.apiResourceBase");
+            String apiWelcomePage = Apl.getStringProperty("apl.apiWelcomeFile");
+
+            response.setContentType("text/html");
+            response.setHeader("X-FRAME-OPTIONS", "SAMEORIGIN");
+            response.setStatus(HttpServletResponse.SC_OK);
+
+            PrintWriter out = response.getWriter();
```
AndrewBoyarsky: Why do we not use `response.sendRedirect("index.html")`?

leonidstr: Redirect doesn't work for it. We should keep an original request path and send it to UI. In this way, UI will find out this path (for example /dashboard) and render page what we expect.

PR #282, opened 2018-12-26
APL-509 - Fix blockchain config updating after block push, add scan reset config

AndrewBoyarsky:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/Genesis.java
```diff
@@ -94,6 +94,7 @@ static void apply() {
         if (genesisAccountsJSON == null) {
             loadGenesisAccountsJSON();
         }
+        CDI.current().select(BlockchainConfig.class).get().reset();
```
yuriy-larin-firstbridge: It seems like that is not needed ???
private static BlockchainConfig blockchainConfig........

oleksiylukin: CDI.current().select... construction is rather for fixing "bad" behavior, but not for "everyday use".

PR #283, opened 2018-12-26
Add Height monitor

AndrewBoyarsky:


oleksiylukin:
I think we have to put all API related classes in pal-api packages and use it in other packages. Please change this.

PR #286, opened 2019-01-12
Fix peer networking

AndrewBoyarsky:


apl-exec/src/main/resources-dev/logback.xml
```diff
@@ -17,10 +17,11 @@
         </logDirectory>
     </define>
     <appender name="file" class="ch.qos.logback.core.rolling.RollingFileAppender">
-        <file>${LOGDIR}/apl.log</file>
+        <!--<file>${LOGDIR}/apl.log</file>-->
```
yuriy-larin-firstbridge: Let's reverte that simplification back, because we are not sure if it works correctly or not.

PR #288, opened 2019-01-15
Added DTO model

maksimkhabenko:


apl-api/src/main/java/com/apollocurrency/aplwallet/api/dto/FbcPropertyDTO.java
```diff
@@ -0,0 +1,13 @@
+package com.apollocurrency.aplwallet.api.dto;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class FbcPropertyDTO {
```
yuriy-larin-firstbridge: Later you can change 'FbcPropertyDTO' name into more suitable. Now it can be like that for the first PR.

apl-api/src/main/java/com/apollocurrency/aplwallet/api/request/BaseRequest.java
```diff
@@ -0,0 +1,9 @@
+package com.apollocurrency.aplwallet.api.request;
+
+/**
+ *
+ * @author al
+ */
+public class BaseRequest {
```
yuriy-larin-firstbridge: Probably it's not needed. You can remove it later.

PR #292, opened 2019-01-22
Feature/apl 517 integrate webui in build

oleksiylukin:
WebUI is being downloaded by Maven build of apl-exec module and unzipped in first run; Some parts of old html directory are needed by APITestServelt of now it is apl-html maven module. It is also unzipped oon first run to the same destination.


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/AplCoreRuntime.java
```diff
@@ -30,6 +31,8 @@
  * @author alukin@gmail.com
  */
 public class AplCoreRuntime {
+    //probably it is temprary solution, we should move WebUI serving out of core 
+    public final static String WEB_UI_DIR="webui";
     private static Logger LOG = LoggerFactory.getLogger(AplCoreRuntime.class);
     private List<AplCore> cores = new ArrayList<>();
```
yuriy-larin-firstbridge: May be better to make it thread safe ? 
private List<AplCore> cores = Collections.synchronizedList( new ArrayList<>() );

oleksiylukin: > May be better to make it thread safe ?
> private List cores = Collections.synchronizedList( new ArrayList<>() );

Though it is not a subject of this PR, I agree. Well, please do it in one of your branches or I'll do it later

apl-exec/src/main/java/com/apollocurrency/aplwallet/apl/exec/webui/WebUiExtractor.java
```diff
@@ -0,0 +1,128 @@
...
+            res=lmd<lmz;
+        }
+        return res;
+    }
+    
+    public void removeDir(String path){
+        try {
+            Files.walk(Paths.get(path))
+                    .map(Path::toFile)
+                    .sorted((o1, o2) -> -o1.compareTo(o2))
```
AndrewBoyarsky: We can use `Comparator.reverOrder()` https://stackoverflow.com/questions/32995559/reverse-a-comparator-in-java-8

PR #311, opened 2019-02-18
Feature/apl 532 prepare for apolloid

oleksiylukin:
Big refactoring of Account and other classes: inner classes moved to outer, a lot of garbage removed.
Well, it is not the end because we have to move to something sane for DB access instead of existing шіт.


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/Shuffling.java
```diff
@@ -522,31 +530,31 @@ public ShufflingParticipant getLastParticipant() {
                 if (!Crypto.isCanonicalPublicKey(publicKey) || !recipientAccounts.add(Account.getId(publicKey))) {
                     // duplicate or invalid recipient public key
                     LOG.debug("Invalid recipient public key " + Convert.toHexString(publicKey));
-                    return new Attachment.ShufflingRecipients(this.id, Convert.EMPTY_BYTES, shufflingStateHash);
+                    return new ShufflingRecipients(this.id, Convert.EMPTY_BYTES, shufflingStateHash);
                 }
             }
             // last participant prepares ShufflingRecipients transaction instead of ShufflingProcessing
-            return new Attachment.ShufflingRecipients(this.id, outputDataList.toArray(new byte[outputDataList.size()][]),
+            return new ShufflingRecipients(this.id, outputDataList.toArray(new byte[outputDataList.size()][]),
```
AndrewBoyarsky: Maybe add suffix `Attachment` to all classes, which implement Attachment interface?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/EntityDbTable.java
```diff
@@ -205,9 +205,8 @@ private T get(Connection con, PreparedStatement pstmt, boolean cache) throws SQL
                 return null;
             }
             T t = null;
-            DbKey dbKey = null;
-            if (doCache) {
-                dbKey = dbKeyFactory.newKey(rs);
+            DbKey dbKey = dbKeyFactory.newKey(rs);
```
AndrewBoyarsky: Db key by default was null, and when `doCache` enabled we set value to db key.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/EntityDbTable.java
```diff
@@ -216,8 +215,12 @@ private T get(Connection con, PreparedStatement pstmt, boolean cache) throws SQL
                     dataSource.getCache(table).put(dbKey, t);
                 }
             }
-            if (rs.next()) {
-                throw new RuntimeException("Multiple records found");
+            //TODO: check this
+            // OL: there are a lot of duplicates really in PUBLIC_KEY table, one with real key and next with null
+            // I do not now why old code does not react on this. So we just log, not thow here
+            if (rs.next() && dbKey!=null) {
```
AndrewBoyarsky: Its normal behavior for derived tables. These tables represents  history logs, so multiple values can be stored at same time. But only one value has flag  `latest=true` and this value used by application. When rollback occurred, some values with `latest=true` can be deleted, so old value with `latest=false` will be updated to `latest=true` and used by application.

PR #352, opened 2019-03-28
Feature/apl 569 shard phased transactions

AndrewBoyarsky:


apl-desktop/src/main/java/com/apollocurrency/aplwallet/apldesktop/DesktopApplication.java
```diff
@@ -723,11 +711,18 @@ public static DbRecoveringUI getInstance() {
                 System.exit(0);
             });
         }
+<<<<<<< HEAD
```
yuriy-larin-firstbridge: Something odd here:
`<<<<<<< HEAD`
`=======`
`>>>>>>> develop`

PR #359, opened 2019-03-31
Feature/apl 582 integrate ref tx

AndrewBoyarsky:


apl-core/pom.xml
```diff
@@ -292,6 +292,14 @@
                     </filesets>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
```
yuriy-larin-firstbridge: I suppose that should be configured in 'top level' POM file for all sub-modules
`<source>11</source>`

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/ReferencedTransactionService.java
```diff
@@ -64,9 +65,9 @@ public boolean hasAllReferencedTransactions(Transaction transaction, int height)
 
     private byte[] getReferencedFullHash(long traansactionId) {
```
yuriy-larin-firstbridge: small typo tr**aa**nsactionId'

PR #366, opened 2019-04-02
retrieving sharded transaction from shard db

yuriy-larin-firstbridge:


apl-core/src/test/java/com/apollocurrency/aplwallet/apl/core/db/ReferencedTransactionServiceTest.java
```diff
@@ -41,13 +42,18 @@
...
+import java.nio.file.Path;
+
 @EnableWeld
 public class ReferencedTransactionServiceTest {
     @Inject
     private JdbiHandleFactory jdbiHandleFactory;
 
     @RegisterExtension
-    static DbExtension extension = new DbExtension();
+    DbExtension extension = new DbExtension(DbTestData.getDbFileProperties(createPath("targetDb").toAbsolutePath().toString()));
```
AndrewBoyarsky: File db is not required here

apl-core/src/test/java/com/apollocurrency/aplwallet/apl/core/phasing/PhasingPollServiceTest.java
```diff
@@ -56,20 +68,15 @@
...
-import java.util.function.Consumer;
-import javax.inject.Inject;
-
 @EnableWeld
 @Execution(ExecutionMode.CONCURRENT)
 public class PhasingPollServiceTest {
 
     @RegisterExtension
-    DbExtension extension = new DbExtension();
+    DbExtension extension = new DbExtension(DbTestData.getDbFileProperties(createPath("targetDb").toAbsolutePath().toString()));
```
AndrewBoyarsky: File db is not required here

apl-core/src/test/java/com/apollocurrency/aplwallet/apl/core/db/ReferencedTransactionServiceTest.java
yuriy-larin-firstbridge: No, it's required here. Otherwise two test methods fail with error:

A file path that is implicitly relative to the current working directory is not allowed in the database URL "jdbc:h2:file:**null**/apl-blockchain-shard-0000001;;MV_STORE=FALSE;CACHE_SIZE=262144". Use an absolute path, ~/name, ./name, or the baseDir setting instead. [90011-196]

apl-core/src/test/java/com/apollocurrency/aplwallet/apl/core/phasing/PhasingPollServiceTest.java
yuriy-larin-firstbridge: No, it's required here. Otherwise two test methods fail with error:

A file path that is implicitly relative to the current working directory is not allowed in the database URL "jdbc:h2:file:**null**/apl-blockchain-shard-0000001;;MV_STORE=FALSE;CACHE_SIZE=262144". Use an absolute path, ~/name, ./name, or the baseDir setting instead. [90011-196]

PR #403, opened 2019-04-23
UPnP fix: use next free port if desired mapping is already busy

oleksiylukin:
This is small fix of UPnP only. New code checks whenever required port is free and if it is free, uses it. If port is already busy it increments number an tries again. This patch fixes at least work of Apollo with UPnP routers. Now multiple Apollo instances do not "fight" for the same port. The same with API.
This patch does not fix P2P transport and announcements because old code uses IP address only. So full fix will be added to "feature/APL-540-node-id" where P2P transport is already refactored.

### DIFF
--- a/apl-utils/src/main/java/com/apollocurrency/aplwallet/apl/util/UPnP.java
+++ b/apl-utils/src/main/java/com/apollocurrency/aplwallet/apl/util/UPnP.java
@@ -14,12 +14,12 @@
  *
  */
 
-/*
+ /*
  * Copyright © 2018-2019 Apollo Foundation
  */
-
 package com.apollocurrency.aplwallet.apl.util;
 
+import java.io.IOException;
 import org.bitlet.weupnp.GatewayDevice;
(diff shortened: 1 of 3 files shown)


apl-utils/src/main/java/com/apollocurrency/aplwallet/apl/util/UPnP.java
```diff
@@ -28,87 +28,128 @@
...
-    public static int TIMEOUT=1500; //1,5 secons
-    private boolean inited=false;
 
+    public static int TIMEOUT = 1500; //1,5 secons
+    private boolean inited = false;
 
-    
 //TODO: Inject with properties    
-    @Inject  
+    @Inject
```
AndrewBoyarsky: Nothing to inject here

oleksiylukin: It's TODO, let it staly

PR #413, opened 2019-05-14
AWW-622 Integration with ETH gas info

leonidstr:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/model/EthGasInfo.java
```diff
@@ -0,0 +1,52 @@
+package com.apollocurrency.aplwallet.apl.exchange.model;
```
oleksiylukin: Why don't you use POJO/JSON mappings?

PR #414, opened 2019-05-15
Added anti-fraud module

0486:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/transaction/TransactionValidator.java
```diff
@@ -67,7 +68,11 @@ public void validate(Transaction transaction) throws AplException.ValidationExce
                 throw new AplException.NotValidException("Transactions of this type must have a valid recipient");
             }
         }
-
+        AntifraudValidator antiFraud = new AntifraudValidator(blockchain.getHeight(), transaction.getSenderId(),
```
oleksiylukin: Please move this constructor call to TransactionValidator constructor;
Make it parameter-less. Set of parameters please move to validate() method of AntifraudValildator

PR #421, opened 2019-05-21
More clean kick-off for wrong application peers

oleksiylukin:
Clean kick-off for peers with wrona application and for peers with wrong chainID
version set to 1.32.12


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/peer/PeerImpl.java
```diff
@@ -704,10 +706,32 @@ void connect(UUID targetChainId) {
                     setState(State.NON_CONNECTED);
                     return;
                 }
+                String app = (String)response.get("application");
+                if(!setApplication(app)){
+                    setState(State.NON_CONNECTED);
```
AndrewBoyarsky: Why we set state and then remove peer with this state?

PR #430, opened 2019-05-24
Feature/aww 1044 implement send money eth/pax

leonidstr:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/DexController.java
```diff
@@ -86,15 +91,28 @@ public DexController() {
...
             @ApiResponse(responseCode = "200", description = "Unexpected error") })
-    public Response getBalances(@QueryParam("eth") String ethAddress, @QueryParam("pax") String paxAddress)
-            throws NotFoundException {
-        return Response.ok(service.getBalances(ethAddress, paxAddress).balanceToJson()).build();
-    }
+    public Response getBalances(@Parameter(description = "Addresses to get balance", required = true) @QueryParam("ethAddresses") List<String> ethAddresses
+        ) throws NotFoundException {
 
+        for (String ethAddress : ethAddresses) {
+            if(ethAddress.length()!=42){
```
AndrewBoyarsky: Can we check it better?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/DexController.java
```diff
@@ -86,15 +91,28 @@ public DexController() {
...
+                return Response.status(Response.Status.OK).entity(incorrect("ethAddress", "Address length is not correct.")).build();
+            }
+        }
+
+        WalletsBalance customerWalletsBalance = service.getBalances(new GetBalancesRequest(ethAddresses));
+
+        try {
+            ObjectMapper mapper = new ObjectMapper();
```
AndrewBoyarsky: Move mapper to constants

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/eth/service/EthereumWalletService.java
```diff
@@ -47,31 +57,105 @@
...
+    }
+
+    /**
+     * Transfer PAX money from account to another one.
+     * @param amountEth
+     * @param gasPrice Gwei
+     * @return String - transaction Hash.
+     */
+    public String transferPax(String passphrase, long accountId, String fromAddress, String toAddress, BigDecimal amountEth, Long gasPrice) throws AplException.ExecutiveProcessException {
+        WalletKeysInfo keyStore = keyStoreService.getWalletKeysInfo(passphrase, accountId);
```
AndrewBoyarsky: Looks like duplicate of prev method.

apl-utils/src/main/java/com/apollocurrency/aplwallet/apl/util/Constants.java
```diff
@@ -127,7 +129,11 @@
 
     public static final String ALPHABET = "0123456789abcdefghijklmnopqrstuvwxyz";
     public static final String ALLOWED_CURRENCY_CODE_LETTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
-    
+
+    //Eth/Pax
+    public static final BigInteger GAS_LIMIT_ETHER_TX = BigInteger.valueOf(21_000);
```
AndrewBoyarsky: Maybe move dex constants to separate class?

leonidstr: It will make sense when we move dex and eth in to another module. Because if we will create new class for dex constants now it makes mess.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/eth/service/EthereumWalletService.java
```diff
@@ -47,31 +57,94 @@
...
+    public BigInteger getPaxBalanceWei(String address){
+        return getTokenBalance(paxContractAddress, address);
+    }
+
+    /**
+     * Get Eth balance.
+     * @param address Eth address
+     * @return account balance in Wei
+     */
+    public BigInteger getEthBalanceWei(String address) {
```
yuriy-larin-firstbridge: That service class can be potentially called not only from Controller, but it can be potentially called from another Service. It would be better to put simplest checks for incoming arguments on most public methods.

I know most of checks are in controller.
What is result of such call
`web3j.ethGetBalance(address, DefaultBlockParameterName.LATEST)` if address is NULL ?

Objects.requireNonNull(...) are very cheap to put into code and they can prevent troubles if it will be called (someday) from another service class. IMHO.

PR #468, opened 2019-06-20
Fix multiple shard processes beginning on same height, fix export/delete/copy of phased transactions, add tests

AndrewBoyarsky:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/helper/BlockDeleteHelper.java
```diff
@@ -48,13 +49,13 @@ public long processOperation(Connection sourceConnect, Connection targetConnect,
...
         paginateResultWrapper.upperBoundColumnValue = upperBoundIdValue;
-
+        Set<Long> excludeDbIds = operationParams.excludeInfo != null ? operationParams.excludeInfo.getNotDeleteDbIds(): Set.of();
         try (PreparedStatement ps = sourceConnect.prepareStatement(sqlToExecuteWithPaging)) {
             do {
                 ps.setLong(1, paginateResultWrapper.lowerBoundColumnValue);
                 ps.setLong(2, paginateResultWrapper.upperBoundColumnValue);
                 ps.setLong(3, operationParams.batchCommitSize);
-            } while (handleResultSet(ps, paginateResultWrapper, sourceConnect, operationParams));
+            } while (handleResultSet(ps, paginateResultWrapper, sourceConnect, operationParams, excludeDbIds));
```
yuriy-larin-firstbridge: It seems like there is no need to pass 'excludeDbIds' as separate params, because that data already passed into inside 'operationParams'. So you can get that Set

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/helper/BlockDeleteHelper.java
```diff
@@ -70,12 +71,12 @@ public long processOperation(Connection sourceConnect, Connection targetConnect,
     }
 
     private boolean handleResultSet(PreparedStatement ps, PaginateResultWrapper paginateResultWrapper,
-                                    Connection sourceConnect, TableOperationParams operationParams)
+                                    Connection sourceConnect, TableOperationParams operationParams, Set<Long> excludeDbIds)
```
yuriy-larin-firstbridge: Here  'TableOperationParams operationParams' still has a 
`'operationParams.excludeInfo.getNotDeleteDbIds()'`
looks like unnecessary memory consumption.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/helper/BlockTransactionInsertHelper.java
```diff
@@ -67,19 +68,19 @@ protected long doStartSelectAndInsert(Connection sourceConnect, Connection targe
...
         }
 
+
         PaginateResultWrapper paginateResultWrapper = new PaginateResultWrapper();
         paginateResultWrapper.lowerBoundColumnValue = lowerBoundIdValue;
         paginateResultWrapper.upperBoundColumnValue = upperBoundIdValue;
-
         long startSelect = System.currentTimeMillis();
-
+        Set<Long> exludeDbIds = operationParams.excludeInfo != null ? operationParams.excludeInfo.getNotCopyDbIds() : Set.of();
```
yuriy-larin-firstbridge: Looks like the same unnecessary memory consumption, don't you think so?

PR #479, opened 2019-06-25
updated PeerInfo and related code

yuriy-larin-firstbridge:
It is not a fix itself, but changes which can help find a problem later


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/peer/endpoint/GetInfo.java
```diff
@@ -66,15 +66,15 @@ protected boolean isChainIdProtected() {
     public JSONStreamAware processRequest(JSONObject req, Peer peer) {
         PeerImpl peerImpl = (PeerImpl)peer;
         PeerInfo pi = mapper.convertValue(req, PeerInfo.class);
-        log.debug("GetInfo - PeerInfo = {}", pi);
+        log.debug("GetInfo - PeerInfo from request = {}", pi);
```
AndrewBoyarsky: request was not logged

PR #509, opened 2019-07-01
Feature/shard restore reset

AndrewBoyarsky:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/TrimService.java
```diff
@@ -60,6 +85,11 @@ public void trimDerivedTables(int height) {
             doTrimDerivedTablesOnBlockchainHeight(height);
             log.debug("Total trim time: " + (System.currentTimeMillis() - startTime));
             dataSource.commit(!inTransaction);
+            if (async) {
+                trimEvent.select(new AnnotationLiteral<Async>() {}).fire(getLastTrimHeight());
```
yuriy-larin-firstbridge: Is it fireAsync() method instead of fire() ??

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/DatabaseManagerImpl.java
```diff
@@ -194,6 +194,16 @@ public TransactionalDataSource createAndAddShard(Long shardId, DbVersion dbVersi
...
     }
 
+    @Override
+    public int closeAllShardDataSources() {
+        int closedDatasources = 0;
+        for (TransactionalDataSource dataSource : connectedShardDataSourceMap.values()) {
+            dataSource.shutdown();
+            closedDatasources++;
+        }
+        return closedDatasources;
```
yuriy-larin-firstbridge: put debug logging here pls. How many DS were closed ?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/TrimService.java
AndrewBoyarsky: No we use such approach, because shard process is not fully asynchronous and we require fully synchronous sharding when restoring trim operations

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/DatabaseManagerImpl.java
AndrewBoyarsky: Its not important for fullReset

PR #523, opened 2019-07-04
updated optionDao to CDI

yuriy-larin-firstbridge:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/model/OptionDAO.java
```diff
@@ -99,8 +100,25 @@ public void deleteAll() {
...
             catch (SQLException e) {
-                LOG.error(e.getMessage());
+                log.error(e.getMessage());
             }
     }
 
+    public boolean exist(String optionKey) {
+        TransactionalDataSource dataSource = databaseManager.getDataSource();
+        try (Connection con = dataSource.getConnection()) {
+            PreparedStatement stmt = con.prepareStatement("SELECT count(*) FROM option WHERE name = ?");
```
AndrewBoyarsky: better to use `select 1 from option` because you want just to check that such record exist

apl-core/src/test/resources/db/data.sql
```diff
@@ -333,4 +333,10 @@ INSERT into shuffling_data
 
 INSERT into trim
 (DB_ID,     HEIGHT,   DONE ) VALUES
-(1    ,      1000,    true);
\ No newline at end of file
+(1    ,      1000,    true);
+
+INSERT into option
+(NAME,     VALUE) VALUES
+('existingKey'    ,      'existingValue'),
```
AndrewBoyarsky: Could you please align data according to column names?

PR #525, opened 2019-07-04
Fix scan for sharded blockchain

AndrewBoyarsky:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardImporter.java
```diff
@@ -0,0 +1,141 @@
...
+        BlockchainProcessor blockchainProcessor = CDI.current().select(BlockchainProcessor.class).get();
+        blockchainProcessor.updateInitialSnapshotBlock();
+        blockchainProcessor.resumeBlockchainDownloading(); // IMPORTANT CALL !!!
+    }
+
+    public void importLastShard(int height) {
+        if (height == 0) {
+            Genesis.apply(false);
+        } else {
+            Shard completedShard = shardDao.getLastCompletedShard();
```
yuriy-larin-firstbridge: NPE can be here if there is not any shard exist

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainProcessorImpl.java
```diff
@@ -1246,21 +1250,30 @@ private void scan(int height, boolean validate, boolean shutdown) {
                 }
                 return;
             }
+            int shardInitialHeight = blockchain.getShardInitialBlock().getHeight();
```
yuriy-larin-firstbridge: I suppose NPE is possible here. When blockchain.getShardInitialBlock() = NULL

AndrewBoyarsky: Should not be null, because at least one block exists in database

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardImporter.java
AndrewBoyarsky: This code should be called only when shard should  be imported and this shard already exists.

PR #550, opened 2019-07-10
Async event when shard is downloaded

oleksiylukin:
Async message on shard download end


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/ShardDownloadPresenceObserver.java
```diff
@@ -75,8 +76,21 @@ public ShardDownloadPresenceObserver(DatabaseManager databaseManager, Blockchain
      *
      * @param shardPresentData shard present data contains downloaded ZIP name
      */
-    public void onShardPresent(@Observes @ShardPresentEvent(ShardPresentEventType.SHARD_PRESENT) ShardPresentData shardPresentData) {
-        shardImporter.importShard(shardPresentData.getFileIdValue(), List.of());
+    public void onShardPresent(@ObservesAsync @ShardPresentEvent(ShardPresentEventType.SHARD_PRESENT) ShardPresentData shardPresentData) {
+        try {
```
AndrewBoyarsky: Why we should wait here?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardImporter.java
```diff
@@ -99,6 +99,11 @@ public void importShard(String fileId, List<String> excludedTables) {
         log.debug("Try unpack file name '{}'", zipInFolder);
         boolean unpackResult = zipComponent.extract(zipInFolder.toString(), csvImporter.getDataExportPath().toString());
         log.debug("Zip is unpacked = {}", unpackResult);
+        if (!unpackResult) {
```
AndrewBoyarsky: Looks like not the best solution

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/ShardDownloadPresenceObserver.java
yuriy-larin-firstbridge: It gives ability to unpack zip successfully. Looks like it works fine now for me locally.

AndrewBoyarsky: Need to find correct solution. Using Thread.sleep - is not acceptable method of synchronization, random aspect is too big

oleksiylukin: Already fixed

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardImporter.java
yuriy-larin-firstbridge: Now it throws exception.

yuriy-larin-firstbridge: Fixed to throw exception

PR #588, opened 2019-08-06
Fix/npe and p2p memory fixes

yuriy-larin-firstbridge:
- fix for NPE on startup with empty db
- added @Singleton to Impl
- attempt to fix stuck PeerWebSocket in memory, but still not fixed completely


oleksiylukin:
This experiment filed, PR introduces trads leaks againt

PR #593, opened 2019-08-14
stop background tasks and discconnect all active peers in Peers::suspend

oleksiylukin:
This patch is needed to completely disable peers while sharding is in progress


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/peer/Peers.java
```diff
@@ -407,12 +409,19 @@ public static void suspend() {
         if(peerHttpServer!=null){
            suspend = peerHttpServer.suspend();
         }
+        TaskDispatcher dispatcher = taskDispatchManager.getDispatcher(BACKGROUND_SERVICE_NAME);
```
yuriy-larin-firstbridge: Do we need 'resume()' method ?
I think we need to add calling both such Peers methods in ShardObserver class.

PR #604, opened 2019-08-27
Fix/attemp optimize trim process

yuriy-larin-firstbridge:
- some changes in trim code
- added more logs in Shuffling, logback.xml for shuffling

### DIFF
--- a/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/observer/TrimObserver.java
+++ b/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/observer/TrimObserver.java
@@ -13,9 +13,9 @@ import com.apollocurrency.aplwallet.apl.util.Constants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.PriorityQueue;
 import java.util.Queue;
 import java.util.concurrent.Executors;
-import java.util.concurrent.PriorityBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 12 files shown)


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/observer/TrimObserver.java
```diff
@@ -30,8 +30,8 @@
     private final TrimService trimService;
     private volatile boolean trimDerivedTables = true;
     private int trimFrequency;
-    private final Object lock = new Object();
-    private final Queue<Integer> trimHeights = new PriorityQueue<>(); // will sort heights from lowest to highest automatically
+//    private final Object lock = new Object();
+    private final Queue<Integer> trimHeights = new PriorityBlockingQueue<>(); // will sort heights from lowest to highest automatically
```
AndrewBoyarsky: Do not use blocking queue here, it will cause incorrect behavior during sharding

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardEngineImpl.java
```diff
@@ -469,8 +469,15 @@ public MigrateState exportCsv(CommandParamInfo paramInfo) {
     }
 
     private int trimDerivedTables(int height) {
-        databaseManager.getDataSource().begin();
+//        databaseManager.getDataSource().begin();
         try {
+            TransactionalDataSource dataSource = databaseManager.getDataSource();
```
AndrewBoyarsky: transaction should be started outside try-catch block

PR #607, opened 2019-09-02
Fix/experiment with shard map

yuriy-larin-firstbridge:
- updated logic for getting transactions from shard db
- updated test data for new logic
- updated unit tests for new logic


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/DbUtils.java
```diff
@@ -171,6 +171,17 @@ public static String limitsClause(int from, int to) {
         }
     }
 
+    public static String limitsClauseInShardDb(int from, int to) {
+        int limit = to >=0 && to >= from && to < Integer.MAX_VALUE ? to - from + 1 : 0;
+        if (limit > 0) { // offset doesn't work in shard db because records were moved
```
AndrewBoyarsky: Why offset doesnt work?

apl-core/src/test/resources/db/shard2-data.sql
```diff
@@ -17,7 +17,7 @@ INSERT INTO PUBLIC.TRANSACTION
...
...d746e4e706daf02b78ba5886f59c204bc2237702d1c2191a6c6b0d3095c9c3d462e4e1cae02f0f53b5e94c2150002b51c553a2e69bc868926235c2fc01ba04b69070324a0c94d9c0d32f65ad4bb475c2b2887800caed2f4023f6510c363a5c4a7da0d8ba7cf85e921990fa7eba87c053ee753157c7541b291483a3f444b0e5d91dcb0f74def9dbe46c910546d0b616ebd9241e7f09aa619cb84b95560307d7e6b07e4fa47c508a621683717485542883203f1f17279b5e93173fa01b19bc707b1ee899bd1118322befed65b6eb28df579d56e61ca6b90abe5408f21544e3e6195ab23876baab07db967de04e815a9395987775acbe57bb7ac8d7366ad62a655bb4598edb4d3d2dce3d326fbeef97b654c686e9abd2c613ea740701a5a4d647e1ebf3bda0fc29fdbb5dfc7dc22842f32e552b0f999076d5f644809ff752224b71fe2f85ad8ac4766d57756d52953bbfb6e6b2134b173bf4995218429371ce3989cd764482396acb05eeaf2e138f38bae9107a9b6db626c6647be5d4a1e6f02f17326700ddeec0b8037671252f0e5c475e06964b6c5a5ff51bc07b494ee84ef5be7d84146f949fe6639409c3fe7550597e45c93ec276721781d9e8677fe4501b583a2b6d96d583c6397c8c5ef14ab6932581d81a8a3518da882fb920dd47c4af25ed755697a7cb181936ae0f21f3c2976f3168202e02fc4b351dcbb7f0c9e5b50a7f1f1d1841dd4de09ca374e3d01fc4fa6cb9271c727a194a2b701ec5e7d882790bb800cc2f86339ad708869ea291105312e302e382000a2c1e47afd4b25035a025091ec3c33ec1992d09e7f3c05875d79e660139220a4'),
```
AndrewBoyarsky: Why do we require such big attachments?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/TransactionDaoImpl.java
```diff
@@ -372,22 +375,38 @@ public synchronized Long getTransactionCount(TransactionalDataSource dataSource,
...
         buf.append("SELECT transaction.* FROM transaction ");
-        createTransactionSelectSql(buf, "transaction.*", accountId, numberOfConfirmations, type, subtype, blockTimestamp, withMessage, phasedOnly, nonPhasedOnly, includeExpiredPrunable, executedOnly, includePrivate, height, prunableExpiration);
-        buf.append(DbUtils.limitsClause(from, to));
+        createTransactionSelectSql(buf, "transaction.*", accountId, numberOfConfirmations, type, subtype,
+                blockTimestamp, withMessage, phasedOnly, nonPhasedOnly, includeExpiredPrunable, executedOnly, includePrivate, height, prunableExpiration);
+        // condition for main or shard db
+        if (dataSource.getDbIdentity().isEmpty()) {
+            buf.append(DbUtils.limitsClause(from, to)); // main db
+        } else {
+            buf.append(DbUtils.limitsClauseInShardDb(from, to)); // shard db
```
AndrewBoyarsky: Is it really necessary?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/DataSourceWrapper.java
```diff
@@ -260,10 +261,10 @@ public void shutdown() {
             stmt.execute("SHUTDOWN COMPACT");
             shutdown = true;
             initialized = false;
+            con.close();
```
AndrewBoyarsky: Looks like con.close may be deleted, because datasource.close will close all connections

apl-core/src/test/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainTest.java
```diff
@@ -924,45 +924,45 @@ void testGetTransactionsWithConfirmation() {
...
     }
 
     @Test
     void testGetTransactionsFromDifferentDataSources() {
         blockchain.setLastBlock(btd.BLOCK_13);
 
-        List<Transaction> transactions = blockchain.getTransactions(txd.TRANSACTION_2.getSenderId(), 0, (byte) -1, (byte) -1, 0, false, false, false, 0, Integer.MAX_VALUE, false, false, true);
-
-        assertEquals(List.of(txd.TRANSACTION_13, txd.TRANSACTION_12, txd.TRANSACTION_11, txd.TRANSACTION_9, txd.TRANSACTION_8, txd.TRANSACTION_7, txd.TRANSACTION_6, txd.TRANSACTION_5, txd.TRANSACTION_4, txd.TRANSACTION_3, txd.TRANSACTION_2, txd.TRANSACTION_1, txd.TRANSACTION_0), transactions);
+        List<Transaction> transactions = blockchain.getTransactions(txd.TRANSACTION_2.getSenderId(), 0, (byte)8, (byte)0, 0, false, false, false, 0, Integer.MAX_VALUE, false, false, true);
```
AndrewBoyarsky: Could you please add additional test case, where bug with retrieving transactions can be reproduced? For now, i see only edited tests.

apl-core/src/test/resources/db/shard2-data.sql
yuriy-larin-firstbridge: It's just a case for type=8 sub-type=0. I took available example data for attachment to prevent NPE for types = 8, 0.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/DbUtils.java
yuriy-larin-firstbridge: Because OFFSET is valid only for full database that was not split to shards.
Split database doesn't have previous offset anymore.
Let's we have two cases :
1.
10 rows in full database. Offset = 5 points to range row numbers = 5...10
2.
We've created two shards like
5 rows in shard  database #1.
and 5 rows in shard  database #2.
Now **offset = 5** doesn't make sense in shard #1 because offset =5 out of db #1 range.

**offset = 5** doesn't make sense in shard #2 too, because offset = 5 points to end of the shard #2. But correct data is within offset = 0...5 for shard #2

https://www.essentialsql.com/what-is-the-differenence-between-top-and-offset-fetch/
**....OFFSET and FETCH are used to return a window of records from a result set.  OFFSET specifies how many rows to skip within the result, and FETCH specifies how many rows from that point forward to return in the result.....**

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/TransactionDaoImpl.java
yuriy-larin-firstbridge: Yes, because offset value is not correct anymore.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/DataSourceWrapper.java
yuriy-larin-firstbridge: may be, it's a minor issue.

apl-core/src/test/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainTest.java
yuriy-larin-firstbridge: Bug is reproducible in previous commit as I wrote in chat.

The main unit test is following.
For given accountId = txd.TRANSACTION_2.getSenderId()
we should fetch all transaction records by type/sub-type = 8, 0
from main + shard1 + shard2 databases, because all those records are split among all those files.

PR #630, opened 2019-09-17
Fix/apl 870 failed release phased tx

AndrewBoyarsky:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/AplDbVersion.java
```diff
@@ -825,7 +825,9 @@ protected int update(int nextUpdate) {
             case 314:
                 apply("ALTER TABLE dex_contract ADD COLUMN IF NOT EXISTS deadline_to_reply INT NOT NULL");
             case 315:
-                return 315;
+                apply("ALTER TABLE phasing_approval_tx ADD COLUMN IF NOT EXISTS latest BOOLEAN NOT NULL DEFAULT TRUE");
```
leonidstr: I think this script is not necessary, because case 310 is not on the production.

PR #633, opened 2019-09-19
Fix/apl 899 mitigate forks

oleksiylukin:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/addons/Demo.java
```diff
@@ -43,9 +43,10 @@ public void init() {
     }
 
     public void onBlockBeforeApply(@Observes @BlockEvent(BlockEventType.BEFORE_BLOCK_APPLY) Block block) {
-        LOG.info("Block " + block.getStringId()
-                + " has been forged by account " + Convert2.rsAccount(block.getGeneratorId()) + " having effective balance of "
-                + Account.getAccount(block.getGeneratorId()).getEffectiveBalanceAPL());
+//Commented out because events are synchronous and this demop call is very slow
```
leonidstr: perhaps it would be better comment this method to avoid call the method at all.

apl-utils/src/main/java/com/apollocurrency/aplwallet/apl/util/cache/CacheConfigurator.java
```diff
@@ -0,0 +1,80 @@
...
+public class CacheConfigurator<K, V> implements CacheConfiguration<K, V> {
+    private String name;
+    private long elementSize;
+    private int cachePriority;
+    private int maxSize = -1;
+    private CacheBuilder cacheBuilder;
+    private Optional<CacheLoader<K, V>> optionalCacheLoader;
+
+    public CacheConfigurator(String name, long elementSize, int cachePriority) {
+        this(name, elementSize, cachePriority, null);
```
yuriy-larin-firstbridge: Andrew added NULL checks here for 'name' values yesterday, as far as I remember.

apl-utils/src/main/java/com/apollocurrency/aplwallet/apl/util/cache/CacheConfigurator.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public CacheConfigurator(String name, long elementSize, int cachePriority, CacheLoader<K, V> cacheLoader) {
+        this.name = name;
```
yuriy-larin-firstbridge: Andrew added NULL checks here for 'name' values yesterday, as far as I remember.

PR #634, opened 2019-09-19
Apl 812 dex atomic swap validation v2

0486:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/service/DexOrderProcessor.java
```diff
@@ -299,7 +299,20 @@ private void processContractsForUserStep2(Long accountId) {
 
     private boolean isContractStep2Valid(ExchangeContract exchangeContract) {
```
leonidstr: We should check how much time left to approved income transfer money(contract.getCounterTransferTxId). 
It should be DEX_CONTRACT_TIME_WAITING_TO_REPLY x 2 since block timestamp.

PR #640, opened 2019-09-21
902 issue: only eth balance matters for gas payments

nemezxxx:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/eth/service/EthereumWalletService.java
```diff
@@ -88,6 +88,18 @@ public EthWalletBalanceInfo balanceInfo(String address){
         return ethWalletBalanceInfo;
     }
 
+    
+     /**
+     * Get Eth balance for both PAX-ETH wallets
```
leonidstr: description is not matching to method's name.

nemezxxx: all necessary alterations have been made. Regard it as done )

PR #643, opened 2019-09-24
APL-834 cache for blockIndex 

alzinchenko:
the inner fields were converted to primitive in BlockIndex;
extracted interface for BlockIndexService;
added cache;
fixed tests;

### DIFF
--- a/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/dao/model/BlockIndex.java
+++ b/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/dao/model/BlockIndex.java
@@ -1,54 +1,24 @@
 package com.apollocurrency.aplwallet.apl.core.db.dao.model;
 
-import java.util.Objects;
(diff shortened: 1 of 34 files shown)


oleksiylukin:
It is highly possible that we also need cache by height and cache of of shard index by height

alzinchenko:
In this case, we need two additional caches by height, one for shards and next one for blocks.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/dao/model/BlockIndex.java
```diff
@@ -1,73 +1,43 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
 
 /**
  * Block global secondary index entity.
  */
+@Getter @Setter
+@EqualsAndHashCode
+@AllArgsConstructor
+@NoArgsConstructor
```
AndrewBoyarsky: Add `@Builder` and remove current builder impl, pls

PR #645, opened 2019-09-24
Apl 911 refund eth pax

leonidstr:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/eth/service/EthereumWalletService.java
```diff
@@ -220,6 +221,15 @@ public String sendApproveTransaction(EthWalletKey ethWalletKey, String spenderAd
         return sendApproveTransaction(ethWalletKey.getCredentials(), spenderAddress, value,  ethGasInfo.getAverageSpeedPrice());
     }
 
+
+    public EthBlock.Block getLastBlock() throws AplException.ExecutiveProcessException {
```
AndrewBoyarsky: Looks like ethBlockNumber do the same

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/service/DexOrderProcessor.java
```diff
@@ -504,6 +518,72 @@ void processCancelOrders(Long accountId) {
         }
     }
 
+    void processExpiredOrders(Long accountId) {
```
AndrewBoyarsky: Looks like this code is almost similar to `processCancelOrders`, i think, that could be better to create general algorithm and customize it for different needs.

AndrewBoyarsky: Add rollback for your cache in `onRollback`

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/eth/contracts/dex.sol
```diff
@@ -0,0 +1,875 @@
...
+    // GETTERS
+    // -----------------------------------------
+
+    function doesUserExist(
+        address user
+    ) external view returns (bool) {
+        return _doesUserExist(user);
+    }
+
+    function getUserFilledDeposits(
```
AndrewBoyarsky: Looks like already withdrawn orders will be also returned

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/eth/service/EthereumWalletService.java
leonidstr: in this case I need block but not a number of the block.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/service/DexOrderProcessor.java
leonidstr: Good idea to refactoring it, but I think it will be better to do after our release.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/eth/contracts/dex.sol
leonidstr: yes, you are right.

PR #662, opened 2019-09-30
Fix/phasing vote table test

alzinchenko:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainImpl.java
```diff
@@ -89,12 +92,10 @@ public BlockchainImpl(BlockDao blockDao, TransactionDao transactionDao, Blockcha
         this.databaseManager = databaseManager;
         this.shardDao = shardDao;
         this.shardRecoveryDao = shardRecoveryDao;
+        lastBlock = new AtomicReference<>();
```
AndrewBoyarsky: Do we really need it?

PR #692, opened 2019-10-16
Add config for order reopening activation on specific height

AndrewBoyarsky:


apl-conf/src/main/resources/conf/chains.json
```diff
@@ -50,6 +50,9 @@
     "prefix": "APL",
     "project": "Apollo",
     "genesisLocation": "conf/data/genesisAccounts.json",
+    "featuresHeightRequirement": {
+      "dexReopenPendingOrdersHeight": 2999999
```
leonidstr: Should we update this value on the a release height than?

PR #698, opened 2019-10-21
Feature/apl 839 corruption issue

yuriy-larin-firstbridge:
- updated P2P code for suspend/resume subsystem to prevent port occupation while sharding
- added public key cache invalidation on rollback
- small shuffling refactoring


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainProcessorImpl.java
```diff
@@ -321,6 +332,10 @@ public BlockchainProcessorImpl(BlockValidator validator, javax.enterprise.event.
         this.taskDispatchManager = taskDispatchManager;
         this.txEvent = txEvent;
 
+        if (propertiesHolder.getBooleanProperty("apl.enablePublicKeyCache")) {
```
AndrewBoyarsky: You can inject specific cache itself instead of cacheManager

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainProcessorImpl.java
```diff
@@ -1069,10 +1086,9 @@ private void accept(Block block, List<Transaction> validPhasedTransactions, List
             }
             log.debug("Total rollback time: {} ms", System.currentTimeMillis() - rollbackStartTime);
             dataSource.clearCache();
-            dataSource.commit(false); // should happen definately, otherwise
-
-        }
-        catch (RuntimeException e) {
+            invalidateAllPublicKeyCache();
```
AndrewBoyarsky: Such invalidation makes publicKeyCache useless.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardEngineImpl.java
```diff
@@ -510,17 +512,34 @@ private void updateRecovery(ShardRecovery recovery, String processedObject) {
...
         if (derivedTable != null) {
             if (!info.isPrunable()) {
+                // not prunable table
                 if (excludedColumns == null) {
+                    // no columns to exclude from export
                     return csvExporter.exportDerivedTable(derivedTable, paramInfo.getSnapshotBlockHeight(), paramInfo.getCommitBatchSize());
                 } else {
-                    return csvExporter.exportDerivedTable(derivedTable, paramInfo.getSnapshotBlockHeight(), paramInfo.getCommitBatchSize(), excludedColumns);
+                    // some column(s) are excluded
+                    if (sortColumn != null && !sortColumn.isEmpty()) {
```
AndrewBoyarsky: I think, that better to override getMaxMinValue (should return min/max order_id) in dexOrderTable and use ordinary exportDerivedTable

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainProcessorImpl.java
yuriy-larin-firstbridge: It's a single way to initialize cache from Manager. There is not another approach to do cache initialization now.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardEngineImpl.java
yuriy-larin-firstbridge: ordinary 'exportDerivedTable' gave incorrect export result due to DB_ID as default sorting column. I switched to new added index by 'height' and use it.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainProcessorImpl.java
yuriy-larin-firstbridge: That was en experiment to check suspicion which was confirmed. Now I commented that out because we need more so sophisticated solution.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardEngineImpl.java
yuriy-larin-firstbridge: I removed 'new index' and reverted that code in latest version.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainProcessorImpl.java
yuriy-larin-firstbridge: Cache is removed

PR #707, opened 2019-10-22
Create index for order table

leonidstr:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/AplDbVersion.java
```diff
@@ -831,8 +831,12 @@ protected int update(int nextUpdate) {
             case 316:
                 apply("CREATE TABLE IF NOT EXISTS mandatory_transaction " +
                         "(db_id IDENTITY, id BIGINT NOT NULL, transaction_bytes VARBINARY NOT NULL, required_tx_hash BINARY(32))");
+
             case 317:
-                return 317;
+                apply("CREATE INDEX IF NOT EXISTS order_tx_id ON dex_offer (id)");
```
AndrewBoyarsky: Maybe rename to `dex_order_id_idx`?

yuriy-larin-firstbridge: The index by ID doesn't help much on derived export, because records are fetched by slq like:
`select * from xxxxx where  db_id> ? and db_id< ? and height <= ? order by db_id limit ?`
in most cases

PR #708, opened 2019-10-22
Feature/integration tests merge develop

maksimkhabenko:


apl-integration-tests/src/test/java/com/apollocurrrency/aplwallet/inttest/model/Wallet.java
```diff
@@ -0,0 +1,54 @@
+package com.apollocurrrency.aplwallet.inttest.model;
+
+public class Wallet {
```
yuriy-larin-firstbridge: I suggest to use lombook annotations here.

maksimkhabenko: Done

PR #723, opened 2019-10-31
Fix/apl 982 scan shards

alzinchenko:
fixed Scan for database with shards


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/helper/CsvImporterImpl.java
```diff
@@ -242,10 +242,10 @@ private long importCsv(String tableName, int batchLimit, boolean cleanTarget,
                 }
                 rsCounter++;
             }
-            dataSource.commit(); // final commit
+            dataSource.commit(false); // final commit
```
AndrewBoyarsky: Maybe add an option to begin transaction, if not started yet?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainProcessorImpl.java
```diff
@@ -1472,7 +1472,9 @@ private void scan(int height, boolean validate, boolean shutdown) {
                             if (validate) {
                                 blockEvent.select(literal(BlockEventType.BLOCK_SCANNED), new AnnotationLiteral<ScanValidate>() {}).fire(currentBlock);
                             } else {
-                                blockEvent.select(literal(BlockEventType.BLOCK_SCANNED)).fire(currentBlock);
+                                if (shardInitialHeight == 0 || currentBlock.getHeight() >= (shardInitialHeight + trimService.getMaxRollback())){
```
AndrewBoyarsky: I think, that its not the best solution. Such event should be triggered independently from sharding configuration

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/helper/CsvImporterImpl.java
yuriy-larin-firstbridge: There is not a problem with starting new transaction, problem with not closing current connection, because external code depends on it and expects it's opened after importer.
I changed that logic because I tried different ways and decided that as most simple one.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/BlockchainProcessorImpl.java
yuriy-larin-firstbridge: If you have better solution, please describe it here. So we can discuss and decide.

PR #729, opened 2019-11-04
Feature/apl perfom test 1

maksimkhabenko:


apl-performance-tests/src/test/scala/com/apollocurrency/PerformanceSimulation.scala
```diff
@@ -0,0 +1,118 @@
+package com.apollocurrency
+
+import com.typesafe.config.ConfigFactory
```
yuriy-larin-firstbridge: Better to have explicit imports without .._  for Scala as well !

PR #731, opened 2019-11-04
APL-943 Fix getting eth nonce.

leonidstr:


leonidstr:
Add to the dev branch  too.

PR #743, opened 2019-11-07
Feature/apl 1044 contract api

AndrewBoyarsky:


apl-crypto/src/main/java/com/apollocurrency/aplwallet/apl/crypto/Convert.java
```diff
@@ -96,6 +96,16 @@ public static long parseUnsignedLong(String number) {
         return Long.parseUnsignedLong(number);
     }
 
+    public static long parseLong(String s) {
```
leonidstr: You can always use parseUnsignedLong()

PR #752, opened 2019-11-13
Exclude prunable_zip_hash from export

AndrewBoyarsky:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/helper/CsvExporterImpl.java
```diff
@@ -252,7 +252,7 @@ public long exportBlockIndex(int targetHeight, int batchLimit) {
      */
     @Override
     public long exportTransactionIndex(int targetHeight, int batchLimit) {
-        MinMaxValue minMaxValue = new MinMaxValue(Long.MIN_VALUE, Long.MAX_VALUE, "transaction_id", 1, targetHeight);
+        MinMaxValue minMaxValue = new MinMaxValue(Long.MIN_VALUE, Long.MAX_VALUE, "transaction_id", 1, targetHeight - 1);
```
yuriy-larin-firstbridge: Why we should do minus 1 now ?

AndrewBoyarsky: We should not export transaction indexes at snapshot block height, since they are already present in the snapshot block. Previous implementation of this method had no such bug.

PR #767, opened 2019-11-19
dublication lock aquiring removed; mostly unused parameter removed

oleksiylukin:
lock that removed was waiting forever in sharding; in other cases this this method is not called directly


oleksiylukin:
At the moment this PR is not ready. Locks now work correctly but rare situation with shard creation is not solved yet

oleksiylukin:
Now shard creation is async on any signal from trim service. This approach works and allows to finish current trim before starting required trims inside of sharding procedure

PR #773, opened 2019-11-20
Do shard synchronously on trimService::init

oleksiylukin:
It is attempt to fix situation when node goes down on the height near sharding height and at the same time avoid dead-locking on cyclic trim execution


oleksiylukin:
another branch takes differrent approcach to shard start

PR #778, opened 2019-11-21
APL-1106 added another reader for quoted text values.

alzinchenko:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/helper/csv/CsvReaderImpl.java
```diff
@@ -251,7 +270,50 @@ private String readValue() throws IOException {
...
+                            break;
+                    }
+                }
+                String s = new String(inputBuffer,
+                        inputBufferStart, inputBufferPos - inputBufferStart - 1);
+                if (!preserveWhitespace) {
+                    s = s.trim();
+                }
+                inputBufferStart = -1;
+                return s;
```
silaev-firstbridge: Just to think about
Is it possible to extract it to a separate method with returning result? 
My concern is that this method is going to be too complex.

PR #779, opened 2019-11-21
Fix/shard skipping

oleksiylukin:


.gitignore
```diff
@@ -23,4 +23,5 @@ nbactions.xml
 **/pom.xml.versionsBackup
 /apl-core/nbproject/
 /apl-integration-tests/allure-results
-/apl-integration-tests/.allure
\ No newline at end of file
+/apl-integration-tests/.allure
+/apl-utils/nbproject/
```
alzinchenko: Is this file required?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/observer/ShardObserver.java
```diff
@@ -45,33 +47,64 @@ public void onTrimDone(@Observes @Sync TrimData trimData) {
...
+    }
+    
+    //TODO: check and debug
+    private boolean isShardAlreadyCreatedOrInProgress(int  lastTrimBlockHeight, HeightConfig currentConfig){
+        Shard shard = shardService.getLastShard();
+        if(shard==null){
+            return false;
+        }        
+        int lastShardHeight = shard.getShardHeight();
+        int nextShardHeight = lastShardHeight+currentConfig.getShardingFrequency();
```
AndrewBoyarsky: You should use shardFrequency calculated in 'isTimeForShard' method.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/observer/ShardObserver.java
```diff
@@ -45,33 +47,64 @@ public void onTrimDone(@Observes @Sync TrimData trimData) {
...
+                    : currentConfig.getShardingFrequency(); // fall back
+        }
+        log.debug("Check shard conditions: ? [{}],  lastTrimBlockHeight = {}, blockchainHeight = {}"
+                + ", shardingFrequency = {} ({})",
+                shardingFrequency != 0
+                        ? lastTrimBlockHeight % shardingFrequency == 0 : "zeroDivision",
+                lastTrimBlockHeight, blockchainHeight,
+                shardingFrequency, blockchainConfig.isJustUpdated());
+        
+        //Q. how much blocks we ould be late? 1/4 of frequiency is OK?
```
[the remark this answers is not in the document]
oleksiylukin: First, we can be late if we miss that trim event and it is quite possible in case of restart. Second, one event only that should trigger all other events is new block acceptance. Triggering sharding from trim and then calling trim from shard is cause of code instability or dead locks. Third, trim is not life-important procedure but we already have too much things bound to it. So we have to prepare for unbinding

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/observer/ShardObserver.java
oleksiylukin: done

apl-utils/src/main/java/com/apollocurrency/aplwallet/apl/util/ModWatcher.java
```diff
@@ -4,26 +4,54 @@
...
+ * You also can get how many full circles your watch did.
+ * Note: you are responsible for calculation of number of hole circles and missed "hours"
  * @author alukin@gmail.com
  */
 public class ModWatcher {
-    int modulos_m;
-    long prevH=0;
-    
-    ModWatcher(int modulos) {
+    long modulos_m;
```
silaev-firstbridge: might it be a good time change it to camel case, like `modulosM` ?

PR #780, opened 2019-11-22
Fix/apl 1062 dex erc20 allowance

andrii-firstbridge:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/eth/service/EthereumWalletService.java
```diff
@@ -242,13 +244,12 @@ public String sendApproveTransaction(EthWalletKey ethWalletKey, String spenderAd
...
      * @param spenderAddress sender address
-     * @param value amount
      * @return tx transaction id
      */
-    private String sendApproveTransaction(Credentials credentials, String spenderAddress, BigInteger value, Long gasPrice){
+    private String sendApproveTransaction(Credentials credentials, String spenderAddress, Long gasPrice) {
         String tx = null;
         try {
-            Function function = approve(spenderAddress, value);
+            Function function = approve(spenderAddress, MAX_POS_INT);
```
AndrewBoyarsky: I think, that amount to approve should be passed by caller code (DexSmartContractService.deposit).

andrii-firstbridge: yep, it makes sense

PR #792, opened 2019-11-26
Feature/apl int test develop

maksimkhabenko:


apl-integration-tests/pom.xml
```diff
@@ -16,6 +16,16 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.29</version>
```
yuriy-larin-firstbridge: You inherit that library from parent pom
```
            <dependency>
                <groupId>org.slf4j</groupId>
                <artifactId>slf4j-api</artifactId>
            </dependency>
```

apl-integration-tests/pom.xml
```diff
@@ -16,6 +16,16 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>1.7.29</version>
```
yuriy-larin-firstbridge: It seems that cab be replaced from parent like:

```
            <dependency>
                <groupId>ch.qos.logback</groupId>
                <artifactId>logback-classic</artifactId>
            </dependency>
```

PR #806, opened 2019-11-29
Feature/apl 952 pass sharding to shell

oleksiylukin:


apl-updater/src/main/java/com/apollocurrency/aplwallet/apl/updater/pdu/AbstractPlatformDependentUpdater.java
```diff
@@ -83,8 +85,17 @@ private void runScript(Path updateDir) {
         }
         try {
             LOG.debug("Starting platform dependent script");
+            PropertiesHolder ph = updaterMediator.getPropertyHolder();
+            
+            boolean isSharding = true; //node's sharding is on by default
+            if(ph!=null){
+               boolean isShardingOff = ph.getBooleanProperty("apl.noshardcreate", false);
```
AndrewBoyarsky: Boolean property is by default 'false', so you can simplify getBooleanProperty call

oleksiylukin: I like obvious code that is easy to read. Defaults is justone of the  worst anti-pattern

PR #813, opened 2019-12-05
removal of extra unused genesis init json files (big)

oleksiylukin:
Removed files are exactly the same, big and unused


oleksiylukin:
already done

PR #839, opened 2019-12-25
Fix/apl tv updates

nemezxxx:
1) including zero intervals
2) dependency of open upon prev close


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/utils/TradingViewUtils.java
```diff
@@ -171,15 +171,12 @@ public static TradingDataOutput getDataForIntervalFromOffers( String fsym, Strin
...
             
-            if (!entryForPeriod.isZero()) {                
-                data.add(entryForPeriod);
-                } else {                
-                if (i==limit-1) {                    
-                    data.add(entryForPeriod);
-                    }
-                }
+            entryForPeriod.open = prevClose;
+            prevClose = entryForPeriod.getClose();
```
AndrewBoyarsky: When current entry is zero, this code will set prevClose to ZERO and next zero/non-zero candlesticks will have zero open price.

AndrewBoyarsky:
Deprecated and replaced by new implementation

PR #840, opened 2019-12-25
[APL-1153] [DO_NOT_MERGE] Remove tx cache

silaev-firstbridge:


silaev-firstbridge:
a part of develop-copy now

PR #846, opened 2020-01-09
Fix/apl 1189 poll result between shard

yuriy-larin-firstbridge:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/derived/DerivedDbTable.java
```diff
@@ -77,6 +77,9 @@ protected DerivedDbTable(String table) {
     @Override
     public void trim(int height) {}
 
+    @Override
+    public void trim(int height, boolean isSharding) {}
```
alzinchenko: I think this is a good place to implement the default behavior of the new trim method, instead of explicit implementation in each 'derived table' class.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/Vote.java
```diff
@@ -60,16 +63,51 @@ public void trim(int height) {
             try (Connection con = databaseManager.getDataSource().getConnection();
                  DbIterator<Poll> polls = Poll.getPollsFinishingAtOrBefore(height, 0, Integer.MAX_VALUE);
                  PreparedStatement pstmt = con.prepareStatement("DELETE FROM vote WHERE poll_id = ?")) {
-                for (Poll poll : polls) {
-                    pstmt.setLong(1, poll.getId());
-                    pstmt.executeUpdate();
-                }
+                commonTrim(height, false, polls, pstmt);
```
AndrewBoyarsky: Did you consider a possibility to just decrease 'Poll.getPollsFinishingAtOrBefore' target height by one or just replace the whole method with new one you created (getPollsFinishingAt, i also suggested to rename it above) instead of supporting two trim variants and increase the complexity of derived tables hierarchy?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/derived/DerivedDbTable.java
yuriy-larin-firstbridge: Done

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/Poll.java
```diff
@@ -135,6 +135,11 @@ public static Poll getPoll(long id) {
         return pollTable.getManyBy(new DbClause.IntClause("finish_height", DbClause.Op.LTE, height), from, to);
     }
 
+    public static DbIterator<Poll> getPollsFinishingAt(int height, int from, int to) {
```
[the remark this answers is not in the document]
yuriy-larin-firstbridge: Method is renamed to getPollsFinishingBelowHeight()

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/Vote.java
yuriy-larin-firstbridge: I wouldn't like to mutate height inside internal method again then it was already mutated before|above. It makes logic not obvious, so I prefer do not do that here.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/Poll.java
```diff
@@ -413,4 +424,14 @@ public boolean isFinished() {
         return partialResult;
     }
 
+    @Override
+    public String toString() {
+        final StringBuffer sb = new StringBuffer("Poll{");
```
silaev-firstbridge: please, consider lombok's @ToString

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/PollOptionResult.java
```diff
@@ -51,4 +51,13 @@ public void add(long vote, long weight) {
         this.weight += weight;
     }
 
+    @Override
+    public String toString() {
+        final StringBuffer sb = new StringBuffer("PollOptionResult{");
```
silaev-firstbridge: same as above mentioned

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/Vote.java
```diff
@@ -156,4 +194,13 @@ public long getVoterId() {
         return voteBytes;
     }
 
+    @Override
+    public String toString() {
```
silaev-firstbridge: same as above mentioned

PR #852, opened 2020-01-14
APL-1213 create new endpoint to get order by id

andrii-firstbridge:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/DexController.java
```diff
@@ -346,6 +346,24 @@ public Response getOffers(  @Parameter(description = "Type of the offer. (BUY =
         ).build();
     }
 
+
+    @GET
+    @Path("/order")
+    @Produces(MediaType.APPLICATION_JSON)
```
AndrewBoyarsky: Maybe better to use plural form for url, because its belong to rest api best practices. Also i suggest to use @PathParam for order id and specify it by using /orders/{orderId}

PR #854, opened 2020-01-15
Add dex order processor config to apl-blockchain.properties

AndrewBoyarsky:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/service/DexOrderProcessor.java
```diff
@@ -88,6 +87,8 @@
     private static final String ETH_SWAP_S1_DETAILS_FORMAT = "secretHash:%s;encryptedSecret:%s";
     private static final String ETH_SWAP_S2_DETAILS_FORMAT = "secretHash:%s";
     private static final int ORDERS_SELECT_SIZE = 30;
+    public static final int DEFAULT_DEX_OFFER_PROCESSOR_DELAY = 3 * 60; // 3 min in seconds
+    public static final int MIN_DEX_OFFER_PROCESSOR_DELAY = 15; // 15 sec
```
andrii-firstbridge: let's move OFFER_PROCESSOR_DELAY to chain.json

AndrewBoyarsky: Chains.json contain consensus config for the whole network, delay for DexOrderProcessor does not affect consesus in any way and can be safely disabled, so that no reason to move it.

PR #856, opened 2020-01-15
APL-1213 add small feature.

andrii-firstbridge:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/DexController.java
```diff
@@ -356,10 +356,14 @@ public Response getOffers(  @Parameter(description = "Type of the offer. (BUY =
         @ApiResponse(responseCode = "200", description = "Unexpected error")})
     public Response getOrder(@PathParam("orderId") String orderIdStr) throws NotFoundException {
         long orderId = Convert.parseLong(orderIdStr);
-
         DexOrder order = service.getOrder(orderId);
 
-        return Response.ok(order.toDto()).build();
+        if (order != null) {
+            DexOrderWithFreezing dexOrderWithFreezing = service.mapToOrdersWithFreezing(order);
```
AndrewBoyarsky: Maybe better to move such logic to DexService.getOrderWithFreezing method for example?

andrii-firstbridge: ok, don't think it so important.

PR #869, opened 2020-01-22
APL-1236 Add functionality to choosing sort by field for method get o…

andrii-firstbridge:
…rders

### DIFF
--- a/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/dao/DexOrderDao.java
+++ b/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/dao/DexOrderDao.java
@@ -38,11 +38,11 @@ public interface DexOrderDao {
             "AND (:status is NULL OR offer.status = :status) " +
             "AND (:offerCur is NULL OR offer.offer_currency = :offerCur) " +
             "AND (:pairCur is NULL OR offer.pair_currency = :pairCur) " +
-            "ORDER BY offer.pair_rate DESC " +
+        "ORDER BY <sortBy> <sortOrder> " +
             "OFFSET :offset LIMIT :limit"
     )
     @RegisterRowMapper(DexOrderMapper.class)
-    List<DexOrder> getOrders(@BindBean DexOrderDBRequest dexOrderDBRequest);
+    List<DexOrder> getOrders(@BindBean DexOrderDBRequest dexOrderDBRequest, @Define("sortBy") String sortBy, @Define("sortOrder") String sortOrder);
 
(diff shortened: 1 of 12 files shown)


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/dao/DexOrderDao.java
```diff
@@ -38,11 +38,11 @@
             "AND (:status is NULL OR offer.status = :status) " +
             "AND (:offerCur is NULL OR offer.offer_currency = :offerCur) " +
             "AND (:pairCur is NULL OR offer.pair_currency = :pairCur) " +
-            "ORDER BY offer.pair_rate DESC " +
+        "ORDER BY <sortBy> <sortOrder> " +
             "OFFSET :offset LIMIT :limit"
     )
     @RegisterRowMapper(DexOrderMapper.class)
-    List<DexOrder> getOrders(@BindBean DexOrderDBRequest dexOrderDBRequest);
+    List<DexOrder> getOrders(@BindBean DexOrderDBRequest dexOrderDBRequest, @Define("sortBy") String sortBy, @Define("sortOrder") String sortOrder);
```
AndrewBoyarsky: Maybe better to use argument factory or map enums directly to sql placeholders?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/model/DBSortOrder.java
```diff
@@ -0,0 +1,25 @@
...
+    ASC("ASC"),
+    DESC("DESC");
+
+    String value;
+
+    DBSortOrder(String value) {
+        this.value = value;
+    }
+
+    public static DBSortOrder ordinal(int ordinal) {
```
AndrewBoyarsky: Resteasy also support static fromString or valueOf methods which accept String parameter and return desired value

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/exchange/model/DexOrderSortBy.java
```diff
@@ -0,0 +1,27 @@
+package com.apollocurrency.aplwallet.apl.exchange.model;
+
+public enum DexOrderSortBy {
+
+    PAIR_RATE("pair_rate"),
+    HEIGHT("db_id");
```
AndrewBoyarsky: Better to rename to DB_ID to lower confusing since height sort will actually perform as following: ORDER BY height ASC/DESC, db_id ASC.

PR #888, opened 2020-02-12
[APL-1238] Move business logic in EntityDbTable to service level

silaev-firstbridge:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/account/service/AccountPublicKeyServiceImpl.java
```diff
@@ -154,13 +154,29 @@ public PublicKey loadPublicKey(DbKey dbKey) {
...
+        PublicKey publicKey = getPublicKey(dbKey, height);
         if (publicKey == null) {
-            publicKey = genesisPublicKeyTable.get(dbKey, height);
+            publicKey = getGenesisPublicKey(dbKey, height);
         }
         return publicKey;
     }
 
+    private PublicKey getGenesisPublicKey(DbKey dbKey, int height) {
+        if (height < 0 || blockChainInfoService.doesNotExceed(height)) {
```
AndrewBoyarsky: genesis public keys are unchangeable, so that such checks are useless

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/account/service/AccountServiceImpl.java
```diff
@@ -117,6 +106,24 @@ public Account getAccount(long id, int height) {
...
 
+    private Account getAccount(DbKey dbKey, int height) {
+        if (height < 0 || blockChainInfoService.doesNotExceed(height)) {
+            return accountTable.get(dbKey);
+        }
+        checkAvailable(height);
+        return accountTable.get(dbKey, height);
+    }
+
+    void checkAvailable(int height) {
```
AndrewBoyarsky: Maybe better to generalize such checks to prevent code duplication and keep blockchain logic consistent?

apl-core/src/test/java/com/apollocurrency/aplwallet/apl/core/db/EntityDbTableTest.java
```diff
@@ -392,146 +394,6 @@ public void testGetManyOnConnectionWithoutCache() {
         );
     }
 
-    @Test
```
AndrewBoyarsky: Too many tests were deleted here, is there a reason for that? Need to replace it or fix to force them to pass successfully or, considering abstract hierarchy, develop special test case for test impl of entityDbTable, such test impl will let us an ability to break tight coupling between this abstract tests and other entities tests.

silaev-firstbridge: I deleted some methods that were used only in tests, so I deleted redundant tests as well. No need to develop something else.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/account/service/AccountPublicKeyServiceImpl.java
silaev-firstbridge: It was done earlier for all entities, so transfer all the logic. Taking into account your point, it may be removed

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/account/service/AccountServiceImpl.java
silaev-firstbridge: I was thinking about while development. A generic method causes more code with Functions that is not quite clear. Better to leave it as I did here.

PR #920, opened 2020-03-10
Fix/apl 1395

alzinchenko:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/AccountController.java
```diff
@@ -378,14 +378,13 @@ public Response getAccountBlocksCount(
...
             description = "Get the block IDs of all blocks forged (generated) by an account in reverse block height order.",
-            security = @SecurityRequirement(name = "admin_api_key"),
             tags = {"accounts"},
             responses = {
                     @ApiResponse(responseCode = "200", description = "Successful execution",
                             content = @Content(mediaType = "application/json",
                                     schema = @Schema(implementation = AccountBlocksCountResponse.class)))
             })
-    @RolesAllowed("admin")
+    @PermitAll
```
AndrewBoyarsky: Need to restrict pagination range set by firstIndex/lastIndex

alzinchenko: fixed

PR #930, opened 2020-03-19
Apl 1428 work with key from file

andrii-firstbridge:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/service/PropertyStorageServiceImpl.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private File getTargetFile(){
+        String configDir = configDirProvider.getConfigDirectory();
+        return Path.of(configDir, EXPORTED_DATA_FILE_NAME).toFile(); // full target file
+    }
+
+    private void createFolderIfNotExist(){
+        String configDir = configDirProvider.getConfigDirectory();
+        File folder = new File(configDir);
```
AndrewBoyarsky: I would suggest Files.createDirectories (no additional checks required)

PR #931, opened 2020-03-24
Fix zero accounts (insert -> delete -> insert for same entity on same height)

AndrewBoyarsky:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/derived/EntityDbTable.java
```diff
@@ -388,6 +396,42 @@ public void insert(T t) {
...
+                                                updatePrevDeleted.setLong(1, prevDbId);
+                                                updatePrevDeleted.executeUpdate();
+                                            } else {
+                                                throw new IllegalStateException("Unable to find previous record for entity " + t + ", inconsistent database state (maybe 'delete' flow is broken)");
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                        } else {
```
yuriy-larin-firstbridge: I suppose it would be good to dump 'context sensitive' information into log before raising exception.
Something like : **entity/table name, current block height, entity PK value, etc...**

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/derived/EntityDbTable.java
```diff
@@ -388,6 +396,42 @@ public void insert(T t) {
                     pstmt.executeUpdate();
                 }
             }
+            if (supportDelete()) {
```
yuriy-larin-firstbridge: More explanations in comments are needed for the code below.

AndrewBoyarsky: Added

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/db/derived/EntityDbTable.java
AndrewBoyarsky: This exception should not be thrown in normal cases, but when thrown - it's easy to understand the real reason. Context data  is not needed, since such error does not depend on it.

PR #932, opened 2020-03-26
Fix/apl 1404 open node health info without admin password

yuriy-larin-firstbridge:
- updated method /control/health to get node's health info _without admin password_
- added method /control/health-full to get full node's info **ONLY WITH admin password**
- added utility method /utils/setlog/level to set logging level dynamically on node


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/UtilsController.java
```diff
@@ -518,4 +523,53 @@ public Response hashByAlgorithm(
...
+                content = @Content(mediaType = "application/json",
+                    schema = @Schema(implementation = SetLogLevelDTO.class)))
+        }
+    )
+    @RolesAllowed("admin")
+    public Response setLoggingLevel(
+        @Parameter(name = "logLevel", description = "Valid log level from available list", required = true,
+            schema = @Schema(implementation = org.slf4j.event.Level.class))
+        @FormParam("logLevel") org.slf4j.event.Level logLevel,
+        @Parameter(description = "The full java package or logger name", required = true) @FormParam("packageName") String packageName,
```
AndrewBoyarsky: Move description to @Schema annotation to make it visible on swagger.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/UtilsController.java
```diff
@@ -518,4 +523,53 @@ public Response hashByAlgorithm(
...
+            @ApiResponse(responseCode = "200", description = "Successful execution",
+                content = @Content(mediaType = "application/json",
+                    schema = @Schema(implementation = SetLogLevelDTO.class)))
+        }
+    )
+    @RolesAllowed("admin")
+    public Response setLoggingLevel(
+        @Parameter(name = "logLevel", description = "Valid log level from available list", required = true,
+            schema = @Schema(implementation = org.slf4j.event.Level.class))
+        @FormParam("logLevel") org.slf4j.event.Level logLevel,
```
AndrewBoyarsky: Better to apply javax.validation on parameters, i guess.

AndrewBoyarsky: org.slf3j.event.Level and ch.qos.logback.classic.Level both used here, i think if we use logback, then better to use level from logback package, isn't it?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/UtilsController.java
yuriy-larin-firstbridge: Updated, but that doesn't show description

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/UtilsController.java
yuriy-larin-firstbridge: Added necessary

yuriy-larin-firstbridge: org.slf3j.event.Level suits much better for showing ENUM on web UI (as dropdown) then ch.qos.logback.classic.Level
That is reason for using two classes from both libraries/

PR #936, opened 2020-04-01
[APL-1451] Remove apl-crypto from apl-api, move converter business logic to service level

silaev-firstbridge:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/service/ServerInfoService.java
```diff
@@ -135,7 +135,11 @@ public ApolloX509Info getX509Info(){
         List<Generator> forgers = Generator.getSortedForgers();
         for (Generator g : forgers) {
             GeneratorInfo gi = new GeneratorInfo();
-            gi.setAccount(Convert.defaultRsAccount(g.getAccountId()));
+            gi.setAccount(
+                Convert.parseAccountId(
+                    Convert.defaultRsAccount(g.getAccountId())
+                )
```
alzinchenko, line 141: Double conversion is unnecessary here,
we can just use gi.setAccount(g.getAccountId())

silaev-firstbridge: done

PR #946, opened 2020-04-15
Feature/apl 1238 repository AskOrder, BidOrder, Trade

silaev-firstbridge:
[APL-1238] AskOrder, BidOrder, Trade


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/trade/model/Event.java
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright © 2018-2020 Apollo Foundation
+ */
+
+package com.apollocurrency.aplwallet.apl.core.trade.model;
+
+/**
+ * @author silaev-firstbridge on 4/14/2020
+ */
+public enum Event {
+    TRADE
```
andrii-firstbridge, line 11: We have about 12 classes which have enum  Event. (Shuffling.Event, CurrencyMint.Event, ....)
So in this case I think it make sense  rename this class on EventTrade for example, to use second part of the name as class name where this event from.

apl-core/src/test/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/AccountEndpointTest.java
```diff
@@ -51,6 +56,7 @@
 import static org.mockito.Mockito.mock;
 
 @EnableWeld
+@Disabled
```
andrii-firstbridge, line 59: Do you want to fix it before merge?

silaev-firstbridge: No, I'm gonna investigate a searching issue in Weld tests after code freeze

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/trade/model/Event.java
silaev-firstbridge: done

PR #957, opened 2020-05-10
Feature/apl 1521 generic response support

AndrewBoyarsky:


apl-utils/src/main/java/com/apollocurrency/aplwallet/apl/util/supervisor/msg/SvBusResponse.java
```diff
@@ -3,19 +3,34 @@
...
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
 public class SvBusResponse extends SvBusMessage {
 
     /**
      * Error code, 0 or null means success
      */
-    public SvBusError error;
+    private SvBusStatus status;
```
andrii-firstbridge, line 29: I think there is no reason to use private access for DTOs. It adds a lot of additional code and annotations. 
But it's question of our team rules.

apl-utils/src/main/java/com/apollocurrency/aplwallet/apl/util/supervisor/client/impl/MessageDispatcherImpl.java
```diff
@@ -103,17 +120,16 @@ void addConnection(URI uri, boolean isDefault) {
...
         ResponseLatch l = waiting.get(rqId);
         if (l != null) {
             res = l.get(RESPONSE_WAIT_TIMEOUT_MS);
         }
         return res;
     }
 
-    public SvBusResponse sendSync(SvBusMessage rq, String path, URI addr) {
-        SvBusResponse res = null;
+    public <T extends SvBusResponse> T sendSync(SvBusMessage rq, String path, URI addr) throws MessageSendingException, SocketTimeoutException {
```
silaev-firstbridge, line 132: Do we really need to ask a user of this mehod to catch throws MessageSendingException, SocketTimeoutException?
Might be better to deal with them inside a method and just throw?

AndrewBoyarsky: There is nothing we can do with those exceptions and such code force the user to handle it to not cause unexpected system fail

silaev: why not just throw this exception inside this method so that not to ask a user to do so?

silaev-firstbridge: why not just throw this exception inside this method so that not to ask a user to do so?

AndrewBoyarsky: Fixed

PR #960, opened 2020-05-12
removed code for dropping indexes after switching to MVStore engine

yuriy-larin-firstbridge:
- removed dropping indexes on shard creation
- added a bit logging

### DIFF
--- a/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardMigrationExecutor.java
+++ b/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardMigrationExecutor.java
@@ -219,6 +219,7 @@ public class ShardMigrationExecutor {
 
     public MigrateState executeAllOperations() {
         stopNetOperations();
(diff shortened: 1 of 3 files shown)


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardMigrationExecutor.java
```diff
@@ -231,7 +232,7 @@ public MigrateState executeAllOperations() {
                 break;
             }
         }
-        log.debug("FINISHED SHARDING '{}'..", state);
+        log.debug("FINISHED SHARDING ----- '{}' in {} ms", state, System.currentTimeMillis() - start);
```
AndrewBoyarsky, line 235: Looks like duplicate of the code below

yuriy-larin-firstbridge: The message is changed only.

PR #961, opened 2020-05-20
OS/Arch updates, add test and serialization/deserialization support

AndrewBoyarsky:


AndrewBoyarsky:
Merged manually, github constantly fails with 500 error

PR #991, opened 2020-07-24
Feature/apl 1614 active generator refactor

yuriy-larin-firstbridge:
Refactored Generator, ActiveGenerators, BlockImpl

### DIFF
--- a/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/runnable/GenerateBlocksThread.java
+++ b/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/runnable/GenerateBlocksThread.java
@@ -5,20 +5,26 @@
 package com.apollocurrency.aplwallet.apl.core.app.runnable;
 
 import javax.enterprise.inject.spi.CDI;
+import java.math.BigInteger;
+import java.security.MessageDigest;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
+import java.util.Map;
+import java.util.Objects;
 
-import com.apollocurrency.aplwallet.apl.core.app.Generator;
 import com.apollocurrency.aplwallet.apl.core.chainid.BlockchainConfig;
+import com.apollocurrency.aplwallet.apl.core.entity.appdata.GeneratorEntity;
 import com.apollocurrency.aplwallet.apl.core.entity.blockchain.Block;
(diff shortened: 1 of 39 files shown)


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/runnable/GenerateBlocksThread.java
```diff
@@ -0,0 +1,158 @@
...
+
+    public void resetSortedForgers() {
+        sortedForgers = null;
+    }
+
+    public void setDelayTime(int delayTime) {
+        this.delayTime = delayTime;
+    }
+
+    private BlockchainProcessor lookupBlockchainProcessor() {
```
AndrewBoyarsky, line 151: Better to avoid lookups in refactored code

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/entity/blockchain/BlockImpl.java
```diff
@@ -228,12 +209,21 @@ public long getPreviousBlockId() {
 
     @Override
     public byte[] getGeneratorPublicKey() {
-        if (generatorPublicKey == null) {
-            generatorPublicKey = lookupAccountService().getPublicKeyByteArray(generatorId);
-        }
+//        if (generatorPublicKey == null) {
```
AndrewBoyarsky, line 212: need to throw here exception if generatorKey was not set

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/app/runnable/GenerateBlocksThread.java
yuriy-larin-firstbridge: Class uses both components like:
private final TransactionProcessor transactionProcessor;
private BlockchainProcessor blockchainProcessor;

but that's not possible inject them both via Constructor, because CDI finds 'cyclic dependency' (between BlockchainProcessor and TransactionProcessor).

So lookup is used to break it.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/entity/blockchain/BlockImpl.java
yuriy-larin-firstbridge: Added throw exception

yuriy-larin-firstbridge: Reverted code back, cause we want : 
check if that field is empty and assign new PubKey value using 'block.getGeneratorId()' method's value.

PR #995, opened 2020-08-03
Feature/apl 1631 finish blockimpl refactor

yuriy-larin-firstbridge:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/entity/blockchain/BlockImpl.java
```diff
@@ -380,69 +342,82 @@ public boolean checkSignature() {
     }
 
     @Override
-    public void setPrevious(Block block, HeightConfig config, Shard lastShard) {
+    public void setPrevious(Block[] threeLatestBlocksArray, HeightConfig config, Shard lastShard, int initialBlockHeight) {
```
andrii-firstbridge, line 345: I think DTO instead of array (threeLatestBlocksArray) will make code more understandable.
For example: 
{
lastBlock,
secondBlock,
thirdBlock
}

AndrewBoyarsky, line 345: I guess that better to move such consensus logic to separate place

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/service/blockchain/BlockchainImpl.java
```diff
@@ -298,7 +301,52 @@ public Block loadBlock(Connection con, ResultSet rs, boolean loadTransactions) {
...
+                parentBlock.setTransactions(Collections.emptyList());
+            }
+        } else if (parentBlock.getTransactions() == null) {
+            parentBlock.setTransactions(Collections.emptyList());
+        }
+        return parentBlock.getTransactions();
+    }
+
+    public JSONObject getJSONObject(Block block) {
+        JSONObject json = new JSONObject();
```
AndrewBoyarsky, line 330: I'd prefer to create separate BlockJsonSerializer for such code snippet

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/entity/blockchain/BlockImpl.java
yuriy-larin-firstbridge: Logic is moved into another component

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/service/blockchain/BlockchainImpl.java
yuriy-larin-firstbridge: Code is moved to another component

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/entity/blockchain/BlockImpl.java
yuriy-larin-firstbridge: Here we need mostly two/three fields like: - timestamp, timeout, baseTarget
so we can optimize that code later after extracting consensus logic

PR #1004, opened 2020-08-13
Bugfix/forging error fix

yuriy-larin-firstbridge:
- added transaction counter into respond
- fixed forging

### DIFF
--- a/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/transaction/messages/AbstractAttachment.java
+++ b/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/transaction/messages/AbstractAttachment.java
@@ -56,7 +56,6 @@ public abstract class AbstractAttachment extends AbstractAppendix implements Att
 
     @Override
     public void validate(Transaction transaction, int blockHeight) throws AplException.ValidationException {
-        this.transactionType = transaction.getType();
         transactionType().validateAttachment(transaction);
     }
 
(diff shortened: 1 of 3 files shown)


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/transaction/messages/AbstractAttachment.java
```diff
@@ -56,6 +56,7 @@ public String getAppendixName() {
 
     @Override
     public void validate(Transaction transaction, int blockHeight) throws AplException.ValidationException {
+        this.transactionType = transaction.getType();
```
andrii-firstbridge, line 59: Why should we set something in side method validate?

yuriy-larin-firstbridge: Because code calls 
public void validate(Transaction transaction...
**transactionType()**.validateAttachment(....

but that is checks :
private TransactionType transactionType() {
        if (**this.transactionType == null**) {
            throw new IllegalStateException("Transaction type was not set");

and then throws IllegalStateException

That is current error on local PC if you try to start forging.

alzinchenko: I guess it's not a good solution.
We need to set the trunsactionType either in the constructor or using the bindTransactionType method

PR #1024, opened 2020-08-31
Feature/apl 1653 trim by soft delete

yuriy-larin-firstbridge:
That works correctly on TN3

### DIFF
--- a/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/service/appdata/TrimService.java
+++ b/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/service/appdata/TrimService.java
@@ -147,11 +147,11 @@ public class TrimService {
                     dbManager.getDataSource().commit(false);
                     int pruningTime = doTrimDerivedTablesOnHeight(trimHeight, false);
                     if (async) {
-                        log.debug("Fire doTrimDerived event height '{}' Async, trimHeight={}", blockchainHeight, trimHeight);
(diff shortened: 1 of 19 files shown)


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/service/appdata/TrimService.java
```diff
@@ -226,19 +226,18 @@ private int doTrimDerivedTablesOnHeight(int height, boolean isSharding) {
...
+//            globalSync.readLock();
             try {
                 long startTime = System.currentTimeMillis();
                 table.prune(pruningTime);
                 table.trim(height, isSharding);
                 dataSource.commit(false);
                 long duration = System.currentTimeMillis() - startTime;
                 // do not log trim duration here, instead go to the logback config and enable trace logs for BasicDbTable class
-                //                    log.trace("Trim of {} took {} ms", table.getName(), duration);
+//                log.debug("Trim of {} took {} ms", table.getName(), duration);
```
AndrewBoyarsky, line 237: Maybe set it to trace back? Too many debug output will be present in the log now

yuriy-larin-firstbridge: That is commented out, so I don't think that is important here.
Usually it's faster to uncomment logging here then configure TRACE logging in XML.

PR #1034, opened 2020-09-10
Adapt sharding for working with mariadb

andrii-firstbridge:


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardService.java
```diff
@@ -142,7 +142,7 @@ public boolean reset(long shardId) {
                     databaseManager.shutdown();
                     FileUtils.deleteFilesByFilter(dirProvider.getDbDir(), (p) -> {
                         Path fileName = p.getFileName();
-                        int shardIndex = fileName.toString().indexOf("-shard-");
+                        int shardIndex = fileName.toString().indexOf("_shard_");
```
yuriy-larin-firstbridge, line 145: What is the reason replacing "dash" by "underscore"? Is there any strong necessity for that?
Probably we are loosing backward compatibility in such case (for old scripts, etc), don't we?

yuriy-larin-firstbridge (approved):

oleksiylukin (approved):

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/shard/ShardService.java
andrii-firstbridge: According naming convention we should use only "A-Z a-z 0-9 _". But in any case we can't save backward compatibility, we had to change db directory according sql-server requirements.

andrii-firstbridge: MariaDb: 
Table and column names are restricted to alphanumeric and underscore only, i.e "A-Z a-z 0-9 _". 
The first character of all table and column names should be an ASCII letter (a-z A-Z).

PR #1043, opened 2020-10-22
Feature/apl 1747 replace dbversion with flyway

andrii-firstbridge:


apl-db-updater/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+  <artifactId>apl-db-updater</artifactId>
+
+  <dependencies>
+
+
+    <!-- Flyway   -->
+    <dependency>
+      <groupId>org.flywaydb</groupId>
+      <artifactId>flyway-core</artifactId>
+      <version>7.0.2</version>
```
yuriy-larin-firstbridge, line 21: Can we move flyway version into **Apollo-box-ext** dependency and use it from there, like org.slf4j, org.projectlombok below ?

alzinchenko (approved):

PR #1055, opened 2020-12-01
Start fixed with no command line parameters; UPnP fix

oleksiylukin:
Start fixed with no command line parameters; Buggy UPnP router behavior 
is fixed (router can not map ports)


oleksiylukin:
wrong branch for merge

PR #1086, opened 2021-02-03
PeerDb changes, removed static methods and initialization

yuriy-larin-firstbridge:
Removed static calls to PeerDb component

### DIFF
--- a/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/peer/PeerLoaderThread.java
+++ b/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/peer/PeerLoaderThread.java
@@ -4,6 +4,7 @@
 package com.apollocurrency.aplwallet.apl.core.peer;
 
 import com.apollocurrency.aplwallet.apl.core.service.appdata.TimeService;
+import lombok.extern.slf4j.Slf4j;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -16,6 +17,7 @@ import java.util.concurrent.Future;
 /**
  * @author alukin@gmail.com
  */
+@Slf4j
 class PeerLoaderThread implements Runnable {
     private static final Logger LOG = LoggerFactory.getLogger(PeerLoaderThread.class);
     private final List<String> defaultPeers;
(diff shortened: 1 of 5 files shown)


alukin (approved):

alzinchenko (approved):

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/peer/PeerLoaderThread.java
```diff
@@ -16,40 +17,48 @@
...
-    public PeerLoaderThread(List<String> defaultPeers, List<Future<String>> unresolvedPeers, TimeService timeService, PeersService peersService) {
+    public PeerLoaderThread(List<String> defaultPeers,
+                            List<Future<String>> unresolvedPeers,
+                            TimeService timeService,
+                            PeersService peersService) {
         this.defaultPeers = defaultPeers;
         this.unresolvedPeers = unresolvedPeers;
         this.timeService = timeService;
         this.peersService = peersService;
+        this.peerDb = peersService.getPeerDb();
```
AndrewBoyarsky, line 38: Maybe better to inject PeerDb?

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/peer/PeerLoaderThread.java
```diff
@@ -16,40 +17,48 @@
...
+            log.error(error);
+            throw new RuntimeException(error);
         }
         final int now = timeService.getEpochTime();
         peersService.wellKnownPeers.forEach((address) -> {
             PeerAddress pa = new PeerAddress(address);
             entries.add(new PeerDb.Entry(pa.getAddrWithPort(), 0, now));
         });
-        if (peersService.usePeersDb) {
+        if (PeersService.usePeersDb) {
```
AndrewBoyarsky, line 55: I guess, that all those options from PeersService (userPeersDb, savePeers, etc) should not be static

oleksiylukin (approved):

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/peer/PeerLoaderThread.java
yuriy-larin-firstbridge: That was first variant, but I changed that later to less params in construcor.

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/peer/PeerLoaderThread.java
yuriy-larin-firstbridge: It seems to me that doesn't matter here.
Most of variables are initialized inside PeersService -> public void init() {......) and later are used as 'read-only' fields.

PR #1092, opened 2021-02-05
Remove updating of peer state on get requests to current blockchain state

AndrewBoyarsky:


oleksiylukin (commented):
It is not obvious when checkBlockchainState() is called before and currentBlockchainState has actual information. Maybe it is better to fix checkBlockchainState()  ?

alzinchenko (approved):

AndrewBoyarsky:
> It is not obvious when checkBlockchainState() is called before and currentBlockchainState has actual information. Maybe it is better to fix checkBlockchainState() ?

I am sure about that change, since every API-level call should not affect node p2p internal state

oleksiylukin (approved):

PR #1172, opened 2021-08-30
fix for account_into

yuriy-larin-firstbridge:
small fix for searching 'account_info' in FTS only


alzinchenko (approved):

apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/service/state/account/impl/AccountInfoServiceImpl.java
```diff
@@ -130,7 +130,7 @@ private StringBuffer createDbIdInRangeFromLuceneData(String luceneQuery) {
         int index = 0;
         try (ResultSet rs = fullTextSearchService.search("public", accountInfoTable.getTableName(), luceneQuery, Integer.MAX_VALUE, 0)) {
             while (rs.next()) {
-                Long DB_ID = rs.getLong(4);
+                Long DB_ID = rs.getLong(5);
```
AndrewBoyarsky, line 133: maybe better to get by column string name to avoid such problems in future

yuriy-larin-firstbridge: Fixed.

PR #1215, opened 2022-04-27
Feature/apl 2149 update certificates

0486:
* Cleanup unused certificates
* Restore UpdaterUtilTest and add new test cases for actual certificates used in updater
* Change certificates for new update with CA replacement and keeping of previous CA cert to allow updates for previous versions
* Update `important update` config with 'update-delay' increase by 6 times (around 10-20 minutes delay)
* Bump version to 1.50.36


AndrewBoyarsky (approved):
Certificates are correct (tested with chain verification and encryption/decryption)

PR #1226, opened 2023-02-27
Vault Wallet file importing error

yuriy-larin-firstbridge:
Fix for Jetty error - IllegalStateException: No multipart config for servlet

### DIFF
--- a/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/KeyStoreController.java
+++ b/apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/KeyStoreController.java
@@ -61,10 +61,6 @@ import java.util.Iterator;
 @Slf4j
 @Path("/keyStore")
 @Singleton
-/*@MultipartConfig(location="./tmp",
-    fileSizeThreshold=64 * 1024, // 64 KB,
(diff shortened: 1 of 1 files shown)


apl-core/src/main/java/com/apollocurrency/aplwallet/apl/core/rest/endpoint/KeyStoreController.java
```diff
@@ -70,12 +75,41 @@ public KeyStoreController(KMSService KMSService, SecureStorageService secureStor
         this.KMSService = KMSService;
         this.secureStorageService = secureStorageService;
         this.maxKeyStoreSize = maxKeyStoreSize;
+        ensureTempUploadDirOnInit();
```
alzinchenko, line 78: I would suggest moving this invocation to the static block of the class and 
remove the condition "if (MULTI_PART_CONFIG == null)" from ensureTempUploadDirOnInit method;

yuriy-larin-firstbridge: fixed

> I would suggest moving this invocation to the static block of the class and remove the condition "if (MULTI_PART_CONFIG == null)" from ensureTempUploadDirOnInit method;

fixed

alzinchenko (approved):

